PR #2, opened 2019-06-19
Feature/add examples

oakca:
Two examples are merged into one.
Now we got a class for Network (nodes and edges).
Network class can create interactive and static maps.
Arrows will be added into maps.


oakca:
@jnnr okey, you can check now. For the plots following:

static: https://ibb.co/s9x1d91
interactive: https://ibb.co/NjLG1rQ

jnnr:
Thanks for your effort! 

* Allow to pass a dictionary that defines colors for different edge types instead of hardcoding. 
* The option to distinguish between elec and heat is nice. Default would be to display heating networks. Can you make the example a pure heating network?
* Is there a possibility to make the background map in folium in greyscale? That makes it easier to see the network.

PR #12, opened 2019-11-13
Features/add osmnx import

jnnr:


jnnr:
Correct styling before merge.

PR #20, opened 2020-01-21
Features/tespy simulation

jnnr:
Adds some facade classes and examples to use tespy for district heating simulations.

* tespy_facades.py: Facades that create tespy components and connections in the background,
* minimal_example.py: A minimal (1 - pipe) example for a network,

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,6 @@
 [flake8]
 exclude = meta/migrations/
-ignore = E265, W503
+ignore = E265, W503, W605
 max-line-length = 100
 per-file-ignores =
     __init__.py : F401
--- a/examples/tespy_simulation/builder_functions.py
+++ b/examples/tespy_simulation/builder_functions.py
@@ -1,20 +1,17 @@
 from tespy.components import (
-    subsystem,
-    splitter,
-    merge,
+    source,
+    sink,
     pipe,
-    valve,
-    heat_exchanger_simple,
 )
+from tespy.connections import connection, bus
 
 from sub_consumer import lin_consum_closed, fork
-
-from tespy.connections import connection
 
 
 def add_producer():
     so = source('source')
     si = sink('sink')
+    new_producer = [so, si]
     return new_producer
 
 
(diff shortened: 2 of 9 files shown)


fwitte:
Hey @jnnr,
I was just scrolling through this. Instead of

> match area independent kA value to U value depending on area

you could easily implement a new equation for the pipe component using ambient temperature, the pipe's and the insulation's dimensions and material as well as the material of the underground to model thermal losses. Just let me know, if you need any assistance :). Also, there is work going on (or has been at least) on a district heating pipe (https://github.com/oemof/tespy/pull/95).
Have a nice sunday!

jnnr:
Hey @fwitte, thanks for your message. Good that you point to that. I will have a look at how to implement new equations in tespy and definitely ask for help if necessary :) Could the same be done (adding an alternative equation) with the Colebrook-White formulation of the pipe's roughness?

fwitte:
Yes, that would be no problem. Colebrook-White is already implemented as standard function: https://github.com/oemof/tespy/blob/dev/tespy/tools/helpers.py#L291-L417. It is used for Reynolds > 2300 and Re*ks/D > 1300.

For other cases of turbulent flow different equations are applied, that (to my knowledge) provide better values for their specific case. However, I checked the differences and the results are very similar. For laminar flow a different equation is applied.

jnnr:
> Yes, that would be no problem. Colebrook-White is already implemented as standard function: https://github.com/oemof/tespy/blob/dev/tespy/tools/helpers.py#L291-L417. It is used for Reynolds > 2300 and Re*ks/D > 1300.

I plotted tespy's lamb function against the formulations I have been using before, and found that they match quite well.

One source of confusion was the definition of the pipe roughness. In my previous calculations, I used the 'absolute roughness', epsilon. Tespy uses the 'equivalent sand grain roughness', ks. It seems that these two are defined differently, but under certain conditions, they can be assumed to be equal. Until now I did not find a good source that clearly explains the differences in the definition. However, this page gives a hint about the difference:

http://daad.wb.tu-harburg.de/tutorial/flood-probability-assessment/hydrodynamics-of-floods/1d-hydrodynamic-models/example/building-up-the-1d-model/equivalent-sand-roughness/

fwitte:
> One source of confusion was the definition of the pipe roughness. In my previous calculations, I used the 'absolute roughness', epsilon. Tespy uses the 'equivalent sand grain roughness', ks.

To my knowledge, the Colebrook-White equation uses the equivalent sand roughness and this parameter is specified in data tables like this https://www.schweizer-fn.de/stroemung/rauhigkeit/rauhigkeit.php. I am not sure, if the equation should be applied to the full range of turbulent flow. The advantage of using different equations lies in the explicit formulation.

fwitte:
Another input from my perspective regarding the looped networks:

It would be really great, if we could get this to work with TESPy. The issue with looped networks is, that the amount of energy balance equations (!!!) at a node of the network (merge or splitter) depends on the flowing direction of the fluid. Therefore, if the flow direction changes within the solution process, the network might get over- or under-determined. I have implemented a workaround using the component node, but this is not working consistently, yet (I think at least). I have another idea, how to tackle this issue, but due to other priorities I did not implement it.

Please let me know, If you are having any trouble with a specific case. I might be able to use that case and test the idea I have in mind.

PR #39, opened 2020-04-08
Setup simulation benchmarks

jnnr:
This PR introduces some minimal examples for thermal networks and their simulation, together with the expected results data. This helps to

* define the problem that are supposed to be solved,
* define the data format in more detail.

In the end, there will be test cases that a simulation model implementation will need to fulfill.

Open TODOs:

* [x] Move the expected results and the two scripts generating it to /tests/_files
* [x] Write a test that runs the tree example and compares it to the expected results.
* [x] Delete the parts for single loop and single pipe- this will be filled with life later.
* [x] The scripts that reproduce the simulation benchmark results have to be adapted to conform with flake8's rules.
* [x] The test that compares the benchmark results with the results from dhnx has to pass.


MaGering:
I've added the calculations for the example of the tree simulation. There are still some issues to discuss:

- The hight, which is needed for the calculation of the hydrostatic pressure difference, is missing as input value. For now I've chosen the latitude as a substitute, because the handling of its data appeared convenient to me.
- The coefficients of local pressure losses are dependent on junction elements and devices. For this example I've made the assumption, that we have tee connectors at the inlet and return where the producer's and consumer's flows meet / split and valves at the consumers inlet as shown in the documentation's figure.
- I've made the calculations with a constant cp. If we drop this assumption we need to take it into account as we calculate the temperature of the mixture at the tee connectors.

examples/simulation/tree/example_calculations/calc_pressure_loss.py
```diff
@@ -0,0 +1,147 @@
...
+g = 9.81            # [m/s2]
+pi = math.pi
+
+# Initialize variables of type dataframe (needed for later calculations)
+v, re, lambda_simp, lambda_adv, dp_diss, dp_loc, dp_loc_tee, dp_loc_valve, dp_hyd, dp = \
+    [pd.DataFrame() for variable in range(10)]
+
+# Adjust mass flows to a dataframe containing all mass flows in correct order
+mass_flow_total = mass_flow.iloc[:, 1:]  # Get mass flows of all consumers
+mass_flow_total['0'] = mass_flow['2'] + mass_flow['3']  # Calculate sum of consumer mass flows as producer mass flow
```
jnnr, line 52: I tried to run it and get a key error here. Maybe you got different input data on your computer?

MaGering: I've just fixed the key error in my recent two commits: 
https://github.com/oemof/DHNx/pull/39/commits/b34e25cba4a6310eb1e2d08f88c165b89a0c53b1
https://github.com/oemof/DHNx/pull/39/commits/0015623c063555b8ef445dd78d31639e7d05ec80

jnnr:
Thanks for your work! I corrected the path in the test and made it possible to compare data with lower precision.

Still missing: Columns and index differ between the default values and DHNx results.

tests/_files/tree_network_sim_expected_results/calc_temp_drop.py
```diff
@@ -0,0 +1,190 @@
...
+temp_drop = temp_drop.iloc[:, 1:]
+# Rename the columns to pipes naming convention
+temp_drop = temp_drop.rename(columns={'0': '1', '1': '2'})
+
+
+def calc_temp_heat_loss(t_in, index):
+    """
+    This function calculates the pipe's outlet temperature
+    out of the inlet temperature due to heat losses
+    :param t_in:
```
jnnr, line 61: Please use numpy-style docstrings

tests/_files/tree_network_sim_expected_results/calc_pressure_loss.py
```diff
@@ -0,0 +1,182 @@
...
+dp_hyd['1'] = - rho * g * abs(forks['m_over_NHN'][0] - consumers['m_over_NHN'].iloc[0]) * \
+              v['1']**0
+dp_hyd['2'] = - rho * g * abs(forks['m_over_NHN'][0] - consumers['m_over_NHN'].iloc[1]) * \
+              v['2']**0
+
+# Calculate total pressure loss
+dp = dp_diss + dp_loc + dp_hyd
+
+# Find the consumer with maximum pressure losses
+max_pressure_losses = max(dp.loc[:, '1':])
```
jnnr, line 120: This seems to only select the pressure loss in the branch of the consumer. In this example it may make no difference to the results, but the whole path from producer to consumer with the highest pressure losses should be chosen.

tests/_files/tree_network_sim_expected_results/calc_temp_drop.py
MaGering: Done in https://github.com/oemof/DHNx/pull/39/commits/4cca64d587b6b43e4a97f5113a45a5967e442937 and https://github.com/oemof/DHNx/pull/39/commits/e560016e6905e6911c6816555fc640fa2bc57b39

tests/_files/tree_network_sim_expected_results/calc_pressure_loss.py
MaGering: So it should be "max_pressure_losses = max(dp.loc[:, '0':])". Is that correct?

PR #42, opened 2020-08-05
Features/simulation models

jnnr:
This introduces a numpy-based model for solving stationary hydraulic and thermal problems. For the beginning, only networks without loops will be considered.

TODO

* [x] Define all material constants in a meaningful place.
* [x] Add a warning that hydrostatic pressure differences due to height difference are not implemented yet.
* [x] Add a warning that looped networks cannot be simulated yet.
* [x] Allow to pass a timeseries for temp_env
* [x] Use zeta values from input data
* [x] check _calculate_pipes_heat_losses - the argument is not used.
* [x] Make sure that docstrings contain the underlying formula
* [x] Fix failing test_setup_simulation
* [x] Document units of results
* [x] Fix pylint and pytest errors
* Checking
  * [x] Compare to benchmark results of the simple tree example 
  * [x] Check if flow direction defined by mass flows is used in all (thermal) calculations, not the defined pipe direction, as the mass flow can go opposite to the defined pipe direction. Update: Opened issue #49.


dhnx/simulation.py
```diff
@@ -10,23 +10,343 @@
...
+            -------
+
+            re : pd.DataFrame
+            """
+            edges_mass_flow = self.results['edges-mass_flow']
+
+            diameter = 1e-3 * self.thermal_network.components.edges['diameter_mm']
+
+            re = 4 * edges_mass_flow.divide(diameter, axis='index') \
+                / (np.pi * self.mu)
```
MaGering, line 129: I suspect that the divide function here is working differently than intended. If so, the mass flows are not divided by the correct diameter.

Here is how it is working right now with the example of the _producer_:

re[0] = 4 * mass_flow_prod[0] / diameter_prod / (pi * mu)
re[1] = 4 * mass_flow_prod[1] / diameter_cons_1 / (pi * mu)
re[2] = 4 * mass_flow_prod[2] / diameter_cons_2 / (pi * mu)

jnnr: Thanks for finding that. Fixed it with f6b8370.

dhnx/simulation.py
```diff
@@ -10,23 +10,440 @@
...
+
+            for i, row in temp_inlet.iterrows():
+
+                mass_flow = self.results['pipes-mass_flow'].loc[i, :].copy()
+
+                temp_difference = np.array(np.dot(row, self.inc_mat)).flatten()
+
+                heat_losses = mass_flow
+
+                pipes_heat_losses[i] = heat_losses.multiply(temp_difference, axis=0)
```
MaGering, line 414: Here the mass flow needs to be multiplied with temp_difference and _self.c_

jnnr: Done in 2197872.

dhnx/simulation.py
```diff
@@ -10,23 +10,440 @@
...
+            ----------
+            re
+
+            Returns
+            -------
+
+            lamb : pd.DataFrame
+            """
+
+            factor_diameter = 1e-3 * self.thermal_network.components.pipes['diameter_mm'] ** -0.14
```
MaGering, line 187:
```suggestion
            factor_diameter = (1e-3 * self.thermal_network.components.pipes['diameter_mm']) ** -0.14
```

Parenthesis are missing around diameter and its unit conversion (1e-3).

dhnx/simulation.py
```diff
@@ -10,23 +10,440 @@
...
+
+            lamb = 0.07 * re **-0.13
+
+            lamb = lamb.multiply(factor_diameter, axis='index')
```
MaGering, line 191: Same issue as with Reynolds calculation. I guess axis must be 'columns' instead of 'index'.

dhnx/simulation.py
jnnr: This should not make a difference.

dhnx/simulation.py
jnnr: Corrected in cc04e1a

jnnr:
Suggestion: temperature_drop -> temp_drop

dhnx/simulation.py
jnnr: Fixed with 4afe05d.

dhnx/simulation.py
```diff
@@ -10,23 +10,445 @@
...
+            pipes_pressure_losses : pd.DataFrame
+                DataFrame with pressure losses values for each timestep.
+            """
+            pipes_mass_flow = self.results['pipes-mass_flow'].copy()
+
+            constant = 8 * lamb / (self.rho * np.pi**2)
+
+            length = self.thermal_network.components.pipes[['from_node', 'to_node', 'length_m']]
+
+            diameter = self.thermal_network.components.pipes[['from_node', 'to_node', 'length_m']]
```
MaGering, line 224:
```suggestion
            diameter = self.thermal_network.components.pipes[['from_node', 'to_node', 'diameter_mm']]
```

dhnx/simulation.py
```diff
@@ -10,23 +10,445 @@
...
+
+            length = length.set_index(['from_node', 'to_node'])['length_m']
+
+            diameter = diameter.set_index(['from_node', 'to_node'])['length_m']
```
MaGering, line 228:
```suggestion
            diameter = diameter.set_index(['from_node', 'to_node'])['diameter_mm']
```

dhnx/simulation.py
```diff
@@ -10,23 +10,445 @@
...
+
+            diameter_5 = 1e-3 * diameter ** 5
```
MaGering, line 230:
```suggestion
            diameter_5 = (1e-3 * diameter) ** 5
```

dhnx/simulation.py
```diff
@@ -10,23 +10,445 @@
...
+
+            pipes_pressure_losses = constant * pipes_mass_flow\
+                .multiply(length, axis='columns')\
+                .divide(diameter_5, axis='columns')
```
MaGering, line 234:
```suggestion
            pipes_pressure_losses = constant * pipes_mass_flow\
                .multiply(pipes_mass_flow, axis='columns')\
                .multiply(length, axis='columns')\
                .divide(diameter_5, axis='columns')
```

It's squared mass flow in this equation. Is there a more elegant way than multiplying it with itself?

jnnr: I squared it before multiplying (2d19cdc .

dhnx/simulation.py
```diff
@@ -10,23 +10,498 @@
...
+
+                return flow_into_node
+
+            nodes_pressure_losses = {}
+
+            for t in self.thermal_network.timeindex:
+
+                pipes_mass_flow = self.results['pipes-mass_flow'].loc[t, :]
+
+                mass_flow_2_over_diameter_4 = pipes_mass_flow.divide(diameter_4)
```
MaGering, line 287:
```suggestion
                pipes_mass_flow = self.results['pipes-mass_flow'].loc[t, :]

                pipes_mass_flow = pipes_mass_flow ** 2

                mass_flow_2_over_diameter_4 = pipes_mass_flow.divide(diameter_4)
```

jnnr: Fixed in 59f3610

jnnr:
@joroeder: We are approaching a point were all results have been checked and have been found equal to the benchmark data. If you find the time, you are invited to have a look at the simulation example and give a review. Still missing: The localized pressure losses, global pressure losses and pump power. The distributed pressure losses in the pipes are already checked.

dhnx/simulation.py
```diff
@@ -10,29 +10,751 @@
...
+            Returns
+            -------
+            nodes_pressure_losses : pd.DataFrame
+                Pressure losses at the nodes.
+            """
+            def _assign_zeta_to_pipes(flow_type):
+
+                def func(row, level):
+
+                    res = zeta.loc[row.index.get_level_values(level).values]
```
MaGering, line 317: I receive a KeyError here: 'Passing list-likes to .loc or [] with any missing labels is no longer supported, see https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#deprecate-loc-reindex-listlike'

I tried fixing it using reindex() instead of loc. Unfortunately it did not work.

MaGering: "res = zeta.reindex(row.index.get_level_values(level).values)" works

jnnr: Thanks for pointing to that. I replaced it and it works. Solved?

MaGering:
```suggestion
                    res = zeta.reindex(row.index.get_level_values(level).values)
```

jnnr: Already did that in 2fed19f

MaGering: Yes, thank you! :)

joroeder (approved):
Hey, I went through all of the code using the example. I think everything is very well structured, and quite easy to follow. Really nice! Altogether, of course, there are always details you can discuss. But for the first version, I think it is totally fine, and a ready-to-use simulation tool!

First, I have some general questions/remarks, which might be good to know for the user or people who want to continue developing, and should be mentioned at some point (maybe in the docs, or in the corresponding docstrings) - please correct me if I got something wrong:

* properties of water are set globally, that means the idea is to take kind of average values, e.g. if you consider a dhs system with 80°C forward and 50°C return, the properties at 65 °C could be a good choice?
* The hydraulic and thermal equations are solved independently. This is possible due to the constant properties
* The flow direction must be known before simulation. Maybe that is already clear, since here tree networks are considered.
* At the moment, there is only the option of one producer.

Some further questions:

* Do the zeta values of the forks for forward and return pipes in such a system differ? So if you join at a tee, or if you split - are there different zeta values? (but maybe this is a question every user needs to think of)
* Is the pressure drop at the heat exchanger of the consumers, and at the heat supply considered?  Or did it simply not occur in the example? Actually, I think this is just a fix delta p added the global_pressure_drop.

Options for improvement/changes:

* Add units in the docstring?
* Instead of providing a temperature drop at the consumers, is it also possible to provide a given return temperature of the heatexchanger? e.g. if you assume that the control variable of the HEX is the return temp, which then can be assumed to be constant.
* Hydraulic results: separately for inlet and return pipes? The way it is, it feels as if everything is thrown into one pot: the pressure drop value for a single pipe could consist of 4 shares: pipe_loss_inlet, pipe_loss_return, fork_inlet_upstream(producer-side), fork_return_upstream(consumers-side) – but maybe that’s fine, I am not sure yet.

Hopefully, I will fine some time to test the simulation tool in a bigger network within our project :)

joroeder:
Of course, the travis issues need to be addressed ...

examples/simulation/single_pipe/sequences/environment-temp_env.csv
```diff
@@ -0,0 +1,4 @@
+snapshot,temp_env
+0,20
+1,20
+2,20
```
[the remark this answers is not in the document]
jnnr: Good idea, let's do it.

examples/simulation/single_pipe/sequences/producers-temp_inlet.csv
```diff
@@ -0,0 +1,4 @@
+snapshot,0
+0,130
+1,130
+2,130
```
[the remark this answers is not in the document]
jnnr: Interesting remarks! Of course, we can take lower numbers. Shall we also write some of this down in the docs? There are two aspects: The problem of too low pressure leading to steam - this be one application of the model: You could set a nominal pressure and check that with the calculated losses it never drops under a given value. The second aspect is that the modeling assumption of constant water properties could be hard to justify in some cases. Could we give some advice on that in the docs?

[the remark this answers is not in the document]
MaGering: According to IFC67 formulation steam pressure of water at 130°C is 2.701 bar. So the water does not evaporate if the nominal pressure is usually PN16 or PN25. Hence evaporation rather does not play a role in the simulation does it?

dhnx/simulation.py
```diff
@@ -10,29 +10,757 @@
...
+        pipes_localized_pressure_losses_return = _calc_loc_pressure_loss_for_flow_type('return')
+
+        pipes_localized_pressure_losses = sum_ignore_none(
+            pipes_localized_pressure_losses_inlet, pipes_localized_pressure_losses_return
+        )
+
+        return pipes_localized_pressure_losses
+
+    def _calculate_global_pressure_losses(self, pipes_pressure_losses):
+
```
MaGering, line 503:
```suggestion
    def _calculate_global_pressure_losses(self, pipes_pressure_losses):
    r"""
        Calculates global pressure losses.

        Parameters
        ----------
        pipes_pressure_losses : pd.DataFrame
            Total pressure losses for every time step and pipe [Pa]

        Returns
        -------
         global_pressure_losses : pd.DataFrame
            Global pressure losses [Pa]
        """
```

dhnx/simulation.py
```diff
@@ -10,29 +10,757 @@
...
+        # Here, we take the path with the maximum pressure losses and assume that the other
+        # consumer's valves are adjusted so that in sum, the pressure losses along all paths are
+        # equal.
+
+        global_pressure_losses = paths_pressure_losses.max(axis=1)
+
+        return global_pressure_losses
+
+    def _calculate_pump_power(self, global_pressure_losses):
+
```
MaGering, line 563:
```suggestion
        r"""
        Calculates the pump power.
        
        .. math::

            P_{el. pump} = \frac{1}{\eta_{el}\eta_{hyd}}\frac{\Delta p }{\rho} \dot{m}

        Parameters
        ----------
        global_pressure_losses : pd.DataFrame
            Global pressure losses [Pa]

        Returns
        -------
         pump_power : pd.Series
            Pump power [W]
        """
```

MaGering (commented):
In this review I've added underlying formula and the parameter's units to the docstrings

dhnx/simulation.py
```diff
@@ -10,29 +10,754 @@
...
+        Parameters
+        ----------
+        temp_node : pd.DataFrame
+            Temperatures at the nodes.
+
+        Returns
+        -------
+        pipes_heat_losses : pd.DataFrame
+            Heat losses in the pipes.
+        """
```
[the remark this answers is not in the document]
jnnr: Done with 6278b11

examples/simulation/single_pipe/sequences/producers-temp_inlet.csv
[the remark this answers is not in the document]
jnnr: Sounds reasonable. If PN - pressure losses is sufficiently higher than 2.7 bar that should be fine.

dhnx/simulation.py
```diff
@@ -10,29 +10,754 @@
...
+
+        pipes_mass_flow = {}
+
+        for t in self.thermal_network.timeindex:
+
+            x, residuals, _, _ = np.linalg.lstsq(
+                self.inc_mat,
+                self.input_data.mass_flow.loc[t, :],
+                rcond=None
+            )
```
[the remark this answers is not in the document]
jnnr: Yes.

dhnx/simulation.py
```diff
@@ -10,29 +10,801 @@
...
+            Known temperatures at producers or consumers.
+
+        direction : +1 or -1
+            For inlet and return flow.
+
+        Returns
+        -------
+        temp_df : pd.DataFrame
+            DataFrame containing temperatures for all nodes.
+        """
```
MaGering, line 669:
```suggestion
                T_{out} = T_{env} + (T_{in} - T_{env}) \cdot exp\{exp_{const} \cdot exp_{var}\} =
                T_{out} = T_{env} + (T_{in} - T_{env}) \cdot exp\{-\frac{U \pi D L}{c \cdot \dot{m}}\}

        Parameters
        ----------
        exponent_constant : np.array
            Constant part of the exponent [kg/s]

        known_temp : pd.DataFrame
            Known temperatures at producers or consumers [°C]

        direction : +1 or -1
            For inlet and return flow [-]

        Returns
        -------
        temp_df : pd.DataFrame
            DataFrame containing temperatures for all nodes [°C]
        """
```

dhnx/simulation.py
```diff
@@ -10,29 +10,801 @@
...
+
+        Parameters
+        ----------
+        temp_inlet : pd.DataFrame
+            Known inlet temperature
+
+        Returns
+        -------
+        temp_return : pd.DataFrame
+            Return temperature with the consumers values set.
```
MaGering, line 757:
```suggestion
            Known inlet temperature [°C]

        Returns
        -------
        temp_return : pd.DataFrame
            Return temperature with the consumers values set [°C]
```

dhnx/simulation.py
```diff
@@ -10,29 +10,801 @@
...
+
+        Parameters
+        ----------
+        temp_node : pd.DataFrame
+            Temperatures at the nodes.
+
+        Returns
+        -------
+        pipes_heat_losses : pd.DataFrame
+            Heat losses in the pipes.
```
MaGering, line 789:
```suggestion
            Temperatures at the nodes [°C]

        Returns
        -------
        pipes_heat_losses : pd.DataFrame
            Heat losses in the pipes [W]
```

MaGering (commented):
In this review I've added the units and checked the underlying formula in the docstrings of the thermal equations

dhnx/simulation.py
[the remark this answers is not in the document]
jnnr: Documented as comment now

dhnx/simulation.py
```diff
@@ -10,29 +10,754 @@
...
+        diameter_5 = (1e-3 * diameter) ** 5
+
+        pipes_pressure_losses = constant * pipes_mass_flow_2\
+            .multiply(length, axis='columns')\
+            .divide(diameter_5, axis='columns')
+
+        # We multiply by the factor of two to represent the pressure losses along inlet
+        # and return flow.
+
+        pipes_pressure_losses *= 2
```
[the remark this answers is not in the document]
jnnr: Documented as comment in the code

Issue #38, opened 2020-04-08
Naming: Edges is a term from graph theory, not energy systems


jnnr:
Proposal: Follow the suggestion that came up in #27 and rename the components edges to pipes. This is a good description and more energy system specific.

PR #43, opened 2020-08-11
Features/update naming

MaGering:
With this PR the naming is adjusted from 'edges' to 'pipes' as suggested in issue https://github.com/oemof/DHNx/issues/38


dhnx/dhn_from_osm.py
```diff
@@ -18,7 +18,7 @@
 from shapely.geometry import LineString
 
 
-def connect_points_to_network(points, nodes, edges):
```
jnnr, line 21: As we talk of nodes here, should we not still talk about edges as well?

dhnx/graph.py
```diff
@@ -29,17 +29,17 @@ def thermal_network_to_nx_graph(thermal_network):
     """
     nx_graph = nx.MultiDiGraph()  # TODO: Check if this line can be removed.
 
-    edge_attr = list(thermal_network.components['edges'].columns)
```
jnnr, line 32: Knowledge about available_components is implicit here.

dhnx/plotting.py
```diff
@@ -40,8 +40,8 @@ class InteractiveMap():
     """
     def __init__(self, thermal_network):
         self.node_data = self.collect_node_data(thermal_network)
```
jnnr, line 42: Again, here it seems to be nodes and edges.

examples/import_osmnx/import_osmnx.py
```diff
@@ -48,23 +48,23 @@
 # get building data
 areas = footprints.area
 
-# get nodes and edges from graph
-nodes, edges = ox.save_load.graph_to_gdfs(graph)
```
jnnr, line 52: It is streets that one gets from osmnx. More abstract: Edges. Stay with edges?

jnnr:
Should add in components.csv's description that pipes have a defined direction.

dhnx/graph.py
jnnr: Suggestion: Let thermal_network as it is, with its concrete components defined in component_attributes.csv. Nodes like consumers, producers are connected by pipes. However, when this function talks about graphs, talk about nodes and edges.

examples/import_osmnx/import_osmnx.py
```diff
@@ -48,23 +48,23 @@
...
-edges.loc[:, ['u', 'v']] = edges.loc[:, ['u', 'v']].replace(replace_ids)
+pipes = pipes.loc[:, ['u', 'v', 'geometry']]
+pipes.loc[:, ['u', 'v']] = pipes.loc[:, ['u', 'v']].replace(replace_ids)
 
 building_midpoints = gpd.GeoDataFrame(footprints.geometry.centroid, columns=['geometry'])
 building_midpoints['x'] = building_midpoints.apply(lambda x: x.geometry.x, 1)
 building_midpoints['y'] = building_midpoints.apply(lambda x: x.geometry.y, 1)
 building_midpoints = building_midpoints[['x', 'y', 'geometry']]
 
-points, splits, edges = dhnx.dhn_from_osm.connect_points_to_network(
```
jnnr, line 66: end_points, forks, edges?

tests/test_errors.py
```diff
@@ -31,34 +31,34 @@
 #         thermal_network.producers['node_id'] = np.float(thermal_network.producers['node_id'])
 #
 #
-# def test_datatype_param_edges():
+# def test_datatype_param_pipes():
 #     with pytest.raises(TypeError):
-#         thermal_network.edges['edge_id'] = np.float(thermal_network.edges['edge_id'])
```
jnnr, line 36: pipe_id has been renamed to id.

tests/test_errors.py
```diff
@@ -31,34 +31,34 @@
 #         thermal_network.producers['node_id'] = np.float(thermal_network.producers['node_id'])
```
jnnr, line 31: Add a note in the code that the assertions that let these tests fail have yet to be implemented.

joroeder:
it seems that sometimes `edges` is sometimes more appropriate, e.g. if you do this logical distinction that edges are pipes before everything (like flow direction, capacity) is defined, then, I am wondering about the optimization part. There, in this sense, you have `pipes` only after the optimization. Before, these are kind of “potential” pipe locations. And in a mashed grid with multiple producers, there can be different load states and flow directions as well. So, I am not sure, if the flow direction should be a must-have for pipes. But, at the moment, I do not have a good idea how to handle. Maybe we need two network attributes: `edges` and `pipes`?

jnnr:
> it seems that sometimes `edges` is sometimes more appropriate, e.g. if you do this logical distinction that edges are pipes before everything (like flow direction, capacity) is defined, then, I am wondering about the optimization part. There, in this sense, you have `pipes` only after the optimization. Before, these are kind of “potential” pipe locations. And in a mashed grid with multiple producers, there can be different load states and flow directions as well. So, I am not sure, if the flow direction should be a must-have for pipes. But, at the moment, I do not have a good idea how to handle. Maybe we need two network attributes: `edges` and `pipes`?

The idea of this PR is to rename edges to pipes because that name is more concrete. At some places in functions that handle networkx graphs I decided that it is more natural to keep talking of edges, as the function is more abstract (see the last commits).

What you propose is to use the two terms, edges and pipes, to discriminate beetween potential pipes and existing pipes. I would like to call them one of the two, edges or pipes. As we discussed this in an issue and there were no objections for quite a while, I would like to stay with 'pipes'.

Then we indicate the distinction of presence or not by setting the attribute 'pipe_type', 'capacity' to None and 0 if it does not exist and some value if it exists. See #44.

joroeder:
> What you propose is to use the two terms, edges and pipes, to discriminate beetween potential pipes and existing pipes. I would like to call them one of the two, edges or pipes. As we discussed this in an issue and there were no objections for quite a while, I would like to stay with 'pipes'.

Okay, let's do it this way.

jnnr (approved):
Thanks for your effort! I reverted the renaming in those cases where the code talks about graphs in a more abstract sense.

PR #44, opened 2020-08-11
Features/integrate solph model builder merge

joroeder:
This PR takes the most important features of #29 and prepares the code for merging.


jnnr:
Thanks for your work! Looks quite impressive. A lot of things seem to have happened since my last visit and everything is much more cleaned up.

**Installation**
I started my review by installing this branch on a fresh environment. No problems at that point. I guess there is a minimum version of oemof.solph (probably 0.4.1). If so, it should be set in setup.py.

**Examples**
Then I tried the examples. For some examples, code and data are contained in their own folder (introduction_example, investment_minimal_example), for the rest that is not the case. If this would be tidied up it would help to gain a quick overview.

Another suggestion (just a small thing): To make names shorter, drop the "_example". As we are in the example folder, it should be clear already. 

investment_input seems to be obsolete. Also, it takes some time to understand which example does what.

The folder would look like this

.
| ---- introduction/
| ---- investment/
| ---- operation/

**Docs**

to be continued

dhnx/input_output.py
```diff
@@ -182,3 +182,21 @@ def load_component_attrs(dir_name, available_components):
         component_attrs[list_name] = df.T.to_dict()
 
     return Dict(component_attrs)
+
+
+def load_invest_options(path):
+
```
jnnr, line 188: Please add a docstring.

dhnx/optimization_modules/add_components.py
```diff
@@ -0,0 +1,329 @@
...
+
+import oemof.solph as solph
+from oemof.tools import economics
+from dhnx.optimization_modules import oemof_heatpipe as oh
+from dhnx.optimization_modules import auxiliary as aux
+
+
+def add_buses(it, labels, nodes, busd):
+    """
+    :param it:  pd.Dataframe containing tabular information for the creation of
```
jnnr, line 23: Please change to numpy format, add explanation and specify return.

dhnx/optimization_modules/add_components.py
```diff
@@ -0,0 +1,329 @@
...
+                    nodes.append(
+                        solph.Source(label=oh.Label(
+                            labels['l_1'], labels['l_2'], labels['l_3'], labels['l_4']),
+                            outputs={busd[l_bus]: solph.Flow(variable_costs=b['shortage costs'])}))
+
+    return nodes, busd
+
+
+def add_sources(on, it, c, labels, gd, nodes, busd):
+
```
jnnr, line 65: Docstring, please :)

dhnx/optimization_modules/add_components.py
```diff
@@ -0,0 +1,329 @@
...
+                        labels['l_1'], labels['l_2'], labels['l_3'], labels['l_4']),
+                    outputs={
+                        busd[(labels['l_1'], cs['label_2'], 'bus',
+                              labels['l_4'])]: solph.Flow(**outflow_args)}))
+
+    return nodes, busd
+
+
+def add_demand(it, labels, gd, series, nodes, busd):
+
```
jnnr, line 126: Docstring. please.

dhnx/optimization_modules/auxiliary.py
```diff
@@ -0,0 +1,40 @@
...
+available from its original location:
+
+SPDX-License-Identifier: MIT
+"""
+
+from oemof.tools import economics
+
+
+def precalc_cost_param(t, q, gd):
+    """
```
jnnr, line 17: Explanation and return would help.

dhnx/optimization_modules/dhs_nodes.py
```diff
@@ -0,0 +1,257 @@
...
+SPDX-License-Identifier: MIT
+"""
+
+
+import oemof.solph as solph
+from dhnx.optimization_modules import oemof_heatpipe as oh, add_components as ac
+
+
+def add_nodes_dhs(opti_network, gd, nodes, busd):
+    """
```
jnnr, line 20: Change to numpy format, please.

dhnx/optimization_modules/dhs_nodes.py
```diff
@@ -0,0 +1,257 @@
...
+    for n, o in opti_network.network.components['forks'].iterrows():
+        d_labels['l_4'] = 'forks-' + str(n)
+        d_labels['l_1'] = 'infrastructure'
+        l_bus = oh.Label(d_labels['l_1'], d_labels['l_2'], d_labels['l_3'],
+                         d_labels['l_4'])
+        bus = solph.Bus(label=l_bus)
+        nodes.append(bus)
+        busd[l_bus] = bus
+
+    # for n, o in opti_network.network.components['producers'].iterrows():
```
jnnr, line 54: Is this still necessary? If yes, a short comment would help.

dhnx/optimization.py
```diff
@@ -33,35 +42,522 @@ def get_results(self):
...
-    r"""
+    """
     Implementation of an invest optimization model using oemof-solph.
+
+    ...
+
+    Attributes
+    ----------
+    settings : dict
+        Dictionary holding the optimisation settings. See .
```
jnnr, line 53: See ...?

dhnx/optimization.py
```diff
@@ -33,35 +42,522 @@ def get_results(self):
 
 
 class OemofInvestOptimizationModel(InvestOptimizationModel):
-    r"""
+    """
     Implementation of an invest optimization model using oemof-solph.
+
+    ...
```
jnnr, line 48: ... ?

dhnx/optimization.py
```diff
@@ -33,35 +42,522 @@ def get_results(self):
...
+          * consumer -> consumer
+          * producer -> producer
+          * producer -> consumer
+          * consumer -> fork
+
+        An error is raised if one of these connection occurs.
+
+        Check 2:
+
+        Check, if the columns of the heat demand series (if given) have the datatype *int*.
```
jnnr, line 115: Why is float not allowed?

examples/optimisation/introduction_example/introduction_example.py
```diff
@@ -0,0 +1,56 @@
...
+# plot network
+static_map = dhnx.plotting.StaticMap(network)
+static_map.draw(background_map=False)
+plt.title('Given network')
+plt.scatter(network.components.consumers['lon'], network.components.consumers['lat'],
+            color='tab:green', label='consumers', zorder=2.5, s=50)
+plt.scatter(network.components.producers['lon'], network.components.producers['lat'],
+            color='tab:red', label='producers', zorder=2.5, s=50)
+plt.scatter(network.components.forks['lon'], network.components.forks['lat'],
+            color='tab:grey', label='forks', zorder=2.5, s=50)
```
jnnr, line 20: Not super urgent, but these lines could be integrated in StaticMap (with a default colormap and the option to pass a custom one).

joroeder:
Thank you for your review so far! I will follow your issues and adapt the code accordingly.

One general thing, I was wondering, if there should be some major change in the structure? Because if so, it would be double work adding docstrings everywhere, and in the next step adapt them again. But if it is fine so far (firstly, it is working quite well ;) , and secondly, some improvements of implementation can be done within the next releases), I would continue adding docstrings. What do you think?

jnnr:
With pleasure! I am only half-way through and am willing to dedicate some time for another review session. Naturally, the more superficial things pop up first and I wanted to write them down. With some more time, I could give some feedback that goes a little deeper. So you are right, it makes sense to wait with the docstrings.

joroeder:
> I started my review by installing this branch on a fresh environment. No problems at that point. I guess there is a minimum version of oemof.solph (probably 0.4.1). If so, it should be set in setup.py.
> 

Yes, I think v0.4.0 is fine.


> ```
> | ---- introduction/
>        | ---- invest_data
>        | ---- tnw_data
>        | ---- introduction.py
> | ---- investment/
>        | ...
> | ---- operation/
>        | ...
> ```

I agree. There is no ```operational``` at the moment - so, take it out?
 

> `network.results.optimization.edges`
> 
>     * contains data on the length, which is not a result, but part of the input data.
> 

Yes, the edges table is extended by the results. In the end, I always appended the results table to the given edges data table to do analysis whatever in qgis ... so it was more convenient to have it directly together.

>     * what is the difference btw. existing and active and invest_status?

* ``active``: only active pipes are considered. If `active==0`, the line element is not considered at all. hopefully this becomes clear from [rtd](https://dhnx.readthedocs.io/en/features-integrate_solph_model_builder_merge/optimization_models.html#edges)
* ``existing``:  this line is not free for optimisation. so a type of heatpipe element must be specifiied and a capacity must be given. And this type must be present in the `invest_data -> network -> pipes.csv` [rtd](https://dhnx.readthedocs.io/en/features-integrate_solph_model_builder_merge/optimization_models.html#edges)
* ``invest_status``: okay, yes, this is not clear ;-) `invest_status` is 1, only for `nonconvex` pipes (so either there is an existing `nonconvex` pipe, or there was an investment in an `nonconvex` pipe). I introduced the parameter to correctly re-calculate the costs and losses in this investment case. The costs and losses are re-calculated for the existing and optimized pipes (in order to have a quick access e.g. to the total heat loss.)

>     * What are these variables?
>       pipe-typ-A.dir-1
>       pipe-typ-A.size-1
>       pipe-typ-A.dir-2
>       pipe-typ-A.size-2
>       pipe-typ-A.size

`pipe-typ-A` -> label of your pipe type from the `pipes.csv`
`dir` -> direction of the oemof-solph heatpipeline Transformer
`size` -> the results of the oemof.solph investment

Generally, you have the option to choose between bi-directional or uni-direction pipes. (see also general settings: [link to rtd](https://dhnx.readthedocs.io/en/features-integrate_solph_model_builder_merge/optimization_models.html#optimization-settings)) Since the flow direction is not known before the optimisation, the model builder creates two pipes in each direction in the case of the setting `bidirectional_pipes=False`). And all these results are given in the results edges.

joroeder:
> 
> 
> With pleasure! I am only half-way through and am willing to dedicate some time for another review session. Naturally, the more superficial things pop up first and I wanted to write them down. With some more time, I could give some feedback that goes a little deeper. So you are right, it makes sense to wait with the docstrings.

Take as much time as you need, (or as you have) ;-)

jnnr:
> I agree. There is no `operational` at the moment - so, take it out?

Yes.

> > `network.results.optimization.edges`
> > ```
> > * contains data on the length, which is not a result, but part of the input data.
> > ```
> 
> Yes, the edges table is extended by the results. In the end, I always appended the results table to the given edges data table to do analysis whatever in qgis ... so it was more convenient to have it directly together.

Your data is explicitly called 'results' and I would strongly suggest to keep the distinction between input data and results very clear. Clearly for analysis and plotting, you often want to combine both. This is still possible in a second step without a lot of effort as the data structure is the same.

dhnx/optimization.py
joroeder: now, ids of consumers, producers, etc. do not need to be of type *int* anymore

joroeder:
Here, I want to provide a short overview of what I have changed and considered from the review so far:

* [ ] ids of consumers, producers, etc. do not need to be of type *int* anymore
* [ ] Add minimum oemof-solph requirement
* [ ] restructure optimisation examples
* [ ] merge dev and rename 'edges' to 'pipes'
* [ ] clean code (remove unused commented code)
* [ ] add missing docstrings
* [ ] ...

to be continued

tests/test_integration.py
```diff
@@ -11,11 +11,11 @@
 
 import os
 
-import pandas as pd
+import helpers
 import networkx as nx
+import pandas as pd
 
 import dhnx
```
jnnr, line 18: This is a relative import of a module in /tests. Was not written very clean which leads to confusion. I made the relative import explicit in https://github.com/oemof/DHNx/pull/42/commits/6d898596eb44f0af16ca991dd67edf7bce15db8b.

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+Scope
+-----
+
+The following questions can be addressed using the `optimize_investment` method
+of the *ThermalNetwork*:
+
+* What is the cost-optimal topology and dimensioning of a DHS piping system?
+  (Given the locations of potential central heat supply plants, the potential
+  locations for the DHS piping system (e.g. street network),
+  and the position of consumers)
```
jnnr, line 29:
```suggestion
* What is the cost-optimal topology and dimensioning of a DHS piping system, 
  given the locations of potential central heat supply plants, the potential
  locations for the DHS piping system (e.g. street network),
  and the position of consumers?
```

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+* In addition to the first question, what is the cost-optimal expansion
+  of a given DHS system.
```
jnnr, line 31:
```suggestion
* In addition to the first question, what is the cost-optimal expansion
  of a given DHS system?957732
```

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+* Is it cost-efficient to build a DHS at all, if there a consumer-wise heat
+  supply alternatives? (Comparison of central and de-central supply strategies)
+* What is the optimal dispatch of the heat producers? (Case, if there are no
```
jnnr, line 34:
```suggestion
* What is the optimal dispatch of the heat producers? (In case there are no
```

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+Overview
+~~~~~~~~
+
+The optimization of a given *ThermalNetwork* is executed by:
+
+.. code-block:: python
+
+    import dhnx
+
+    tnw = dhnx.network.ThermalNetwork()
```
jnnr, line 67:
```suggestion
    tnw = dhnx.network.ThermalNetwork()
    
```

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+
+* **id**: see :ref:`Thermal Network <thermal_network_label>`
+* **from_node**: see :ref:`Thermal Network <thermal_network_label>`
+* **to_node**: see :ref:`Thermal Network <thermal_network_label>`
+* **length**: see :ref:`Thermal Network <thermal_network_label>`
+
+The following optional attributes are introduced by the optimization module:
+
+* **existing**: Binary indicating an existing pipe. If there is no column
+  *existing* given, all Edges are free for optimization.
```
jnnr, line 156:
```suggestion
  *existing* given, all Pipes are free for optimization.
```

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+
+    tree
+    ├── pipes.csv                       # (required)
+    ├── consumers.csv                   # (required)
+    ├── forks.csv                       # (required)
+    ├── producers.csv                   # (required)
+    └── sequences                       # (optional)
+        └── consumers-heat_flow.csv
+
+The attributes, which are required, and which are optional with respect
```
jnnr, line 133: It is very helpful to know which attributes are necessary in particular for optimisation. Good to have these tables. As an idea for a cleaner structure, we could in a future release define all requirements in the respective specificatoins of dhnx/components_attrs. I will open an issue.

jnnr:
About the dictionary `settgs`: Why not rather pass keyword arguments and set default values in the function? As a sketch:

in network.py, you could pass keyword arguments

```
def optimize_investment(self, invest_options, **kwargs):

    oemof_opti_model = setup_optimise_investment(
        self, invest_options, **kwargs
    )

```
in optimization.py, you could set the default arguments

```
setup_optimise_investment(thermal_network, invest_options, heat_demand='scalar', etc.)
```

With this approach, there would be an error if you pass a setting that does not exist.

jnnr:
In the examples, there seems to be an inconsistent naming, or am I wrong? Maybe it is not too severe, but consistency always helps

* invest_data
* invest_options

docs/optimization_models.rst
jnnr: Just for readability, maybe add a space here?

docs/optimization_models.rst
jnnr: The issue is #52. We can do this later.

docs/optimization_models.rst
jnnr: Renaming not completed yet. Might occur at other places as well.

docs/optimization_models.rst
```diff
@@ -2,4 +2,619 @@
...
+
+Producers
+'''''''''
+
+.. csv-table::
+   :header-rows: 1
+   :file: _static/opti_producers.csv
+   :align: center
+
+The following optional attributes must be given in every *ThermalNetwork*:
```
jnnr, line 205: I would rather not repeat the attributes of the ThermalNetwork. Repeated information means extra maintenance work. Just show the extra attibutes required for optimization models. In the beginning of the section, this could be clarified: Here we only present the optimizatoin specific attributes.

docs/optimization_models.rst
joroeder: Yes, I think this would be very helpful in future. Thanks for opening the issue.

jnnr:
When I checkout the branch locally and try to run the tests, I get: 

```
pytest gives the error:

    from oemof.solph import helpers
E   ImportError: cannot import name 'helpers'
```

It seems that I have the right oemof.solph v0.4.1 installed though.

docs/optimization_models.rst
joroeder: Yes, you are right. I cut the tables to the additional attributes of the optimisation approach ...

docs/optimization_models.rst
```diff
@@ -1,5 +1,610 @@
...
+The following figure provides an overview of the input data:
+
+.. 	figure:: _static/DHNx_Input_data.svg
+   :width: 100 %
+   :alt: optimization_input_data.svg
+   :align: center
+
+   Fig. 1: Optimisation Input Data
+
+The structure of the input data might look a bit confusing at the beginning, but provides a lot of
```
jnnr, line 96: :) a bit informal, but I like the honesty.

docs/optimization_models.rst
```diff
@@ -1,5 +1,610 @@
...
+options for building up complex district heating models. There are two groups of data:
+geo-referenced data and parameters for the investment optimisation (with might include other
+oemof-solph related data). All data needs to be provided in csv files. This means the geo-referenced
```
jnnr, line 99: This sentence contradicts the schematic above and even more directly, the previous sentence. Proposal: There are two groups of data: Firstly, data that describes the components and the connectivity of the network, required by the ThermalNetwork class. Secondly, data that is necessary for the investment optimization.

docs/optimization_models.rst
```diff
@@ -1,5 +1,610 @@
...
+Input Data
+~~~~~~~~~~
+
+In this section, it is firstly revised, what input data is exactly necessary
+from the *ThemalNetwork* class, and then explained, what data needs to be
+provided as `invest_options` and as `settings`.
+
+The following figure provides an overview of the input data:
+
+.. 	figure:: _static/DHNx_Input_data.svg
```
jnnr, line 89: Even if I like the picture very much, it seems that some information does not apply in general or repeats specifications of ThermalNetwork. For the moment, I would be ok to leave it like it is to proceed with the release - maybe we can address this later?

joroeder: yeah, I already did some small updates in order to bring it more into line with the current state, however, as you say, there are still some issues. e.g. the sequences are also part of the thermal network. Other sequences are not ... I will update a more simplified version to give a rough overview.

Issue #40, opened 2020-08-04
Move graph and osm import functions to input_output module


jnnr:
There is an example for importing street networks from osm. The functionality is not included in main code base. Should be moved to the input_output.py-module. ~~The same applies to the conversion function from and to nx.graph.~~ Update: The functions for converting a ThermalNetwork to an nx.DiGraph is already in a separate module called graph.py

PR #47, opened 2020-09-28
Clean up OSM import

jnnr:
This PR addresses #40, moving the functionality of the osm example to an importer class.

Open TODOs

* [x] For consumers: include lat, lon
* [x] For forks: remove x, y 
~~* [ ] Edges are not divided when nearest points are introduced~~ Update: Separate issue #48


joroeder (commented):
This PR does not change the functionality, right? Because the osm import is a very important tool and has a lot of potential. However, eg the processing (connection the buildings) needs some adaption to make it fully compatible with the optimization and simulation part. But if this PR does not change the functionality, there is nothing against merging. It is still very valuable to have it already included.

jnnr:
You are right. This does not change optimization or simulation models. There are still some steps missing to use it in a workflow with optimization and simulation. This would be the next steps - using the tools and find out how the different parts can interplay with another.

joroeder:
I am wondering, is geopandas so far a requirement? Yes, but only for the osmnx import, right? And besides osmnx, if we want to have geojson Importer Classes we need also geopandas, right? But for the simulation and optimisation part, there is no need for geopandas at the moment, right?
I think the osmnx import is very powerful, and maybe everybody, who is doing something with district heating systems in python, needs at some point geopandas. So, in the end, we will have geopandas anyway, right? What do you think?

jnnr:
Geopandas is in "extras_require" in setup.py. It is needed for osm import but not for optimisation and simulation. As soon as someone writes a geojson importer, we can add a try import except to the input_output module.

jnnr:
Regarding the remaining problem of the edges that are not divided when new nodes are introduced: The problem is here

https://github.com/oemof/DHNx/blob/dev/dhnx/dhn_from_osm.py#L82

New edges are just the old edges combined with the new, connecting edges. One step needs to be added: Cutting the edges in two at the new nodes. Can someone help?

joroeder:
> 
> 
> Geopandas is in "extras_require" in setup.py. It is needed for osm import but not for optimisation and simulation. As soon as someone writes a geojson importer, we can add a try import except to the input_output module.

Sounds good. However, at the moment it is in ``install_requires`` - I am confused ;-)

> New edges are just the old edges combined with the new, connecting edges. One step needs to be added: Cutting the edges in two at the new nodes. Can someone help?

Do you want to have this in the first release? I think, of course, this would be nice. But on the other hand, we can always add this features later. The import so far shows at least the principle workflow, and for the second release, this can be done unhurriedly.

jnnr:
> > Geopandas is in "extras_require" in setup.py. It is needed for osm import but not for optimisation and simulation. As soon as someone writes a geojson importer, we can add a try import except to the input_output module.
> 
> Sounds good. However, at the moment it is in `install_requires` - I am confused ;-)

You are right. osmnx is in extras_require. Geopandas is a dependency of osmnx. So geopandas could be removed from install_requires, it seems. Should test if that works though.
 
> > New edges are just the old edges combined with the new, connecting edges. One step needs to be added: Cutting the edges in two at the new nodes. Can someone help?
> 
> Do you want to have this in the first release? I think, of course, this would be nice. But on the other hand, we can always add this features later. The import so far shows at least the principle workflow, and for the second release, this can be done unhurriedly.

In my view, it is a bug. But the dev has the bug as well, so we could still merge this PR if everything else looks good.

jnnr:
I moved geopandas to extras require. With this change, the users can decide for themselves to install it. If they do not install it, some examples will not work, but the import error is caught and a message is printed.

jnnr:
In my view this is ready. The linters' warnings are dealt with in the branch of the simulation models, so they should be resolved when we first merge #42.

PR #58, opened 2020-10-26
Fixes/clean up component attrs

### LINKED ISSUE #46: Naming: Inconsistent use of 'temp' and 'temperature'
To shorten names, temp seems preferable over temperature. Files and variable names using the long version should be renamed.

### LINKED ISSUE #57: ThermalNetwork: components.csv and component_attrs needs clean-up
Some of the information given in components and components_attrs is not up-to-date any more. Before the release, we need a review and clean up.

In components.csv

* [x] Delete ThermalSubnetwork - does not exist yet. Deleted
* [x] TransferStation same
* [x] ThermalStorage same
* [x] environment - is it used? Yes
* [x] Input/Output: Is this used? Not really. Can consider this when integrating optimization and simulation. OK for now.

In components_attrs:

* [x] unit rendered as dot. Render as hyphen.
* [x] node_type necessary? Yes.
* [x] rename node_type to component_type and add to requirements
* [x] Some variables have different variable names in different places (with and without units):

    ThermalNetwork vs NumpySimulationModel
    length  length_m
    diameter  diameter_mm
    heat_transfer_coeff  heat_transfer_coefficient_W/mK
    roughness  roughness_mm

* [x] In type, mention that these can be either float/series

   mass_flow
   temp_drop
   temp-env

jnnr:
Fixes #57 and fixes #46 as well.

I processed all the steps listed in #57. On the way, I also removed all entries on the component_type in the input data with the idea to let this be set as default. However, `CSVImporter` did not yet set the default attributes. Until now, this only happened when adding components one by one using the `add` method. I introduced a new method `set_defaults` that is called in the `CSVImporter`.


joroeder (approved):
I think looks good. The `length` attribute name is now also consistent now to the optimisation part.

Issue #49, opened 2020-10-08
Simulation: Calculations rely on defined pipe direction


jnnr:
It is not clear if flow direction defined by mass flows is used in all (thermal) calculations, not the defined pipe direction. This has to be checked.

The mass flow can go opposite to the defined pipe direction, in particular in the case of looped networks. They are not implemented at the moment, but even with tree networks, this can introduce an error if the user does not define all pipes in the direction of the mass flow.

jnnr:
It should be rather easy to test this: Define a simple tree with pipe direction reversed (from leaf node towards producer).

jnnr:
This is a bug. The results depend on the pipe directions defined in the input data.

jnnr:
This is resolved by #60

PR #60, opened 2020-11-02
Fix bug swapping pipe direction

jnnr:
Addresses #49.

TODO:
* [x] Review
* [x] Add section to whats new


joroeder (approved):
From my side it is fine! The test is also a good thing.

PR #63, opened 2020-11-16
Fix/no bidirectional consumer connection

joroeder:
Bug: If the setting `bidirectional_pipes=True` is selected, also the connecting line to the building was created as bidirectional.

This only matters if a source or a storage is added at the consumers.

This PR fixes the issue.

### DIFF
--- a/dhnx/optimization_add_components.py
+++ b/dhnx/optimization_add_components.py
@@ -347,7 +347,7 @@ def add_storage(it, labels, nodes, busd):
     return nodes
 
 
-def add_heatpipes(it, labels, bidi, q, b_in, b_out, nodes):
+def add_heatpipes(it, labels, bidirectional, q, b_in, b_out, nodes):
     """
     Adds *HeatPipeline* objects with *Investment* attribute to the list of oemof.solph components.
 
@@ -387,7 +387,7 @@ def add_heatpipes(it, labels, bidi, q, b_in, b_out, nodes):
         # bidirectional heatpipelines yes or no
         flow_bi_args = {
             'bidirectional': True, 'min': -1}\
-            if bidi else {}
+            if bidirectional else {}
 
         nodes.append(oh.HeatPipeline(
             label=oh.Label(labels['l_1'], labels['l_2'],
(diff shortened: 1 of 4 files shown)


dhnx/optimization_add_components.py
```diff
@@ -357,8 +357,8 @@ def add_heatpipes(it, labels, gd, q, b_in, b_out, nodes):
         Table of *Heatpipeline* attributes of the district heating grid
     labels : dict
         Dictonary containing specifications for label-tuple
-    gd : dict
-        Settings of the investment optimisation of the ThermalNetwork
+    bidi : bool
```
jnnr, line 360:
```suggestion
    bidirectional : bool
```

jnnr: A bit longer, but easier to understand. Could consider giving more telling names to 'q' and 'b_in' (bus_in, e.g.)

jnnr (approved):
I tested the 3 examples and they run. I did not test if the bug was solved as I am not too deep into the optmization model and do not know what to look for. If this solves your problem, I am fine with it. 

However, the different handling for different cases is now somehow hard-coded within the if-else statements. I do not understand the code that much that I could come up with an idea, but this may be worth thinking.

To make the code more readable and maintainable, it would certainly help to refactor it in a way that avoids the nested if else statements in optimization_dhs_nodes.py

joroeder:
Thank you for reviewing!

> 
> However, the different handling for different cases is now somehow hard-coded within the if-else statements. I do not understand the code that much that I could come up with an idea, but this may be worth thinking.

The reason is that it should be possible to have the option of using bidirectional pipelines, which means a bidirectional flow in solph, or not. If not, two pipelines in each direction are created. Maybe, this if else statment is not needed, because in solph a bidirecitonal flow directly sets the `min` attribute to -1. This would be a new issue for further improvment.

> 
> To make the code more readable and maintainable, it would certainly help to refactor it in a way that avoids the nested if else statements in optimization_dhs_nodes.py

Yes, thats a good idea. I refactored the `q`.

PR #64, opened 2020-11-16
Fix/osmnx import

### LINKED ISSUE #48: OSM import: Edges are not divided when new nodes are introduced
New edges are just the old edges combined with the new, connecting edges. One step needs to be added: Cutting the edges in two at the new nodes. 

https://github.com/oemof/DHNx/blob/dev/dhnx/dhn_from_osm.py#L82

joroeder:
The osmnx_import example is not running.

On thing is that there was a wrong attribute name `distance`  instead `dist`.

The other issue, I don't know, I commented the lines. The principal import works now, however, likely not properly at the current state.
Could you have a look at it @jnnr ?

### DIFF
--- a/dhnx/input_output.py
+++ b/dhnx/input_output.py
@@ -403,8 +403,6 @@ class OSMNetworkImporter(NetworkImporter):
 
         footprints = self.download_footprints()
 
-        # footprints = ox.project_gdf(footprints)
-
         component_dfs = self.process(graph, footprints)
 
         self.add_component_data_to_network(component_dfs)
--- a/dhnx/input_output.py
+++ b/dhnx/input_output.py
@@ -401,7 +401,7 @@ class OSMNetworkImporter(NetworkImporter):
 
         footprints = self.download_footprints()
 
-        footprints = ox.project_gdf(footprints)
+        # footprints = ox.project_gdf(footprints)
 
         component_dfs = self.process(graph, footprints)
 
--- a/dhnx/input_output.py
+++ b/dhnx/input_output.py
@@ -206,8 +206,10 @@ class OSMNetworkImporter(NetworkImporter):
 
         print('Downloading footprints...')
 
-        footprints = ox.footprints.footprints_from_point(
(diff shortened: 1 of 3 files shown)


jnnr:
> I think it would be good to solve this before continuing the actual issue with the division of edges ..

So this does not fix issue #48? Then it should not say so. Can you instead state the issue you are solving with this PR?

joroeder:
> 
> So this does not fix issue #48? 

This PR was indendet to do so, and it is not finshed yet ;-)

However, I see your point. It makes sense to use a separte PR for that. I will update the PR describtion.

joroeder:
the `distance` attribute changed in osmnx v0.13.0:

https://github.com/gboeing/osmnx/blob/master/CHANGELOG.md#0130-2020-05-25

joroeder (commented):
Some general questions and comments on the osmnx import

dhnx/input_output.py
```diff
@@ -401,7 +401,7 @@ def load(self):
 
         footprints = self.download_footprints()
 
-        footprints = ox.project_gdf(footprints)
+        # footprints = ox.project_gdf(footprints)
```
joroeder, line 404: Are the footprints already projected in line 215? What I don't understand, what is the original crs, and what the projected?

joroeder:
> 
> 
> Some general questions and comments on the osmnx import

Please ignore this comment.

joroeder:
Okay, I fixed the issues regarding the not-running osmnx-import example, which means an update according to the actual osmnx version (0.16.1). I added this version requirement to the `extras_require` of `setup.py`.

The other issues (like #48) of the osmnx import are not solved with this PR.

joroeder:
@jnnr Since this branch is for review for more than 4 months now ;-) and it contains only small fixes by updating the API of the osmnx package, I would merge this branch if I don't hear anything by the end of next week.

jnnr (approved):
Tested successfully.

Issue #54, opened 2020-10-22
Integrate/Improve geometry pre-processing


joroeder:
This issue is for discussing the integration of functionalties and tools for the pre-processing of the geometry in [https://github.com/quarree100/gistools](https://github.com/quarree100/gistools)

This could also help to solve the bug of #48

joroeder:
The gistools were integrated and are already published with v0.0.2.

PR #65, opened 2020-11-16
Features/move gistools to dhnx

joroeder:
Related issue: #54 

* [x] Specify module description
* [x] Complete all docstrings
* [x] Update Documentation
* [x] Update WhatsNew
* [x] Add illustrative example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ _build
 jupyter_notebooks
 */import_export_plot/data_csv_output/
 */import_osmnx/data/
+cache/
--- a/dhnx/gistools/__init__.py
+++ b/dhnx/gistools/__init__.py
@@ -1,2 +1,2 @@
-from dhnx.gistools import connect_points
-from dhnx.gistools import geometry_operations
+from . import connect_points
+from . import geometry_operations
--- a/dhnx/gistools/connect_points.py
+++ b/dhnx/gistools/connect_points.py
@@ -126,15 +126,22 @@ def cut_line_at_points(line_str, point_list):
     return lines
 
 
-def create_object_connections(points, lines, tol_distance=2):
+def create_object_connections(points, lines, tol_distance=1):
     """Connects points to a line network.
 
     Generally, the nearest point of the next line is used as connection the point.
-    Depending on the geometry, there are 3 ways, the connection is created:
-    - lot food point is outside the line
-    - lot food point is on the line
-    - lot food point is on the line, but the line ending is closer than `tol_distance`.
-      In that case the end of line is used as connection point.
+    Depending on the geometry, there are 3 options, the connection is created:
+    - nearest point is line ending => the connection line starts from this line ending
+    - nearest point is on the next line:
+        a) line endings are outside the tolerance => line is split and the nearest point
+           is used as connection point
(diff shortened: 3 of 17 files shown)


jnettels:
Just FYI: While not an error, I have encountered an unclosed file ResourceWarning when running the example "import_osm_invest.py". It seems to be related to osmnx and I have opened an issue there: https://github.com/gboeing/osmnx/issues/607

jnettels:
I hope you don't mind, I added a tiny commit with additional info for configuring the cbc solver. This took me a while to get to work and should be useful here.
I would also like to properly set up the logging module for the "gistools" code (as noted here #69), but that does not need to happen in this PR.

joroeder:
Yes, thank you. On the one hand, it is something solver-specific. But on the other hand, you are right, it might be very helpful and also shows how to apply and hand these options to the solver ...

jnettels:
Sorry, I rely on Spyder's code style check, but flake8 is even pickier than that...

joroeder:
Okay, this one is now finished from my side, I gave the last finishing touches from my side after standing still for so long. This is almost a shame, as this PR contains such valuable steps for setting up the network from your individual gis data.

I also included the most important hints (e.g. to the example showing the whole model chain from osm to optimisation results) in a docs chapter: [https://dhnx.readthedocs.io/en/features-move_gistools_to_dhnx/geometry_preparation.html](https://dhnx.readthedocs.io/en/features-move_gistools_to_dhnx/geometry_preparation.html)

@jnnr It would be great if you could have a look at it. Since it is mainly additional code, I do not expect that there are major flaws concerning the existing functionalities of the package. I would appreciate if we could merge this one soon ...

joroeder:
Thank you for your response!

> 
> On a more general level: It would be good to describe the PR a bit better. What problem does it solve and how? Which problems does it not solve?

Yes, sorry, I agree ;-) Please see below. I thought that we have talked about that, but nontheless, you are right.

> 
> How do the functions introduced here stand in relation to the one in the module dhn_from_osm, containing a function `connect_points_to_network`? Is it replaced by the new functions?

Yes, this would be the next step and fix the bug that the lines are not split in the function `connect_points_to_network`. But therefore, I found it convenient to first fix the existing osm import (#64) and provide our existing approach (this PR) in the DHNx package, we are currently using in our projects. And then, we should use a separate Issue/PR for replacing the old function.

> 
> Will the new functions be integrated in the OSMNetworkImporter in the future? The Importer class has the benefit that it wraps the different steps that are performed in the example `import_osm_invest` and hides the details of the implementation, which makes it easier to use and to maintain.

Yes, this could/should be integrated in future (I will open a new issue for that). However, the design philosophy differs and needs some discussion and work to integrate it properly (This is also the reason for not doing it directly with this PR. For example, it was not intended to wrap everything, because the users should be able to use their own buildings layer and eg only the network from osm. Also, the proceeding differs at many points in detail. I also think that by using and already providing the current approach, we can figure out the best way to improve the NetworkImporter – but all this can be discussed in future issues/PR)

jnettels (approved):
Looks good to me. The use cases I could test work as expected. Sorry for taking so long.

PR #77, opened 2021-07-09
Restructure / refactor optimization modules

### LINKED ISSUE #67: Missing API documentation of heatpipeline in Readthedocs
The API of the heatpipeline is not added to Readthedocs. The module is missing in the `API Documentation` section itself. And, additionally, it could be added to the main part of documentation for better understanding, what is the background of the optimization model.

joroeder:
This PR restructures the optimization modules and puts them into ```dhnx.optimization``.

For the user, (at least for all examples), this does change anything as the network optimization is implemented as method of the *ThermalNetwork* (`ThermalNetwork.optimize_investment()`). It is all under the hood.

For the simulation part, there are no changes at all, it only addresses the optimization part.

 * [x] refactor modules
 * [x] check and fix sphinx API documentation build (fixes #67 )
   ~~Add CI again~~ > done by #78 
 * [x] Update WhatsNew


joroeder:
@jnettels @jnnr If somebody could have a short look on that one, e.g. check the examples, would be really nice :)

As a re-structuring sometimes can imply conflicts, however, if there are errors at restructuring, normally, the imports do not work, which I could not encounter here.

Maybe, a short comment, on what do you think about the here proposed putting all modules concerned with optimization in a folder, would be great. I also think, if the development continues, we can also re-think the modules structure time by time. For now, this seems to make sense to me, so I could add further modules for pre- and post-processing for the optimization in submodules in that folder.

PR #110, opened 2022-12-13
New method 'boundary' for creating the building connection point

### LINKED ISSUE #107: Access violation error in unary_union() with shapely and pygeos installed
The function create_object_connections() uses the shapely unary_union() operation.

There seems to be a conflict between shapely and pygeos, which both use 'geos' internally, if both are installed.
This can cause ``OSError exception: access violation reading 0xFFFFFFFFFFFFFFFF`` when unary_union() is used.

With shapely 1.8.0 and pygeos 0.12.0 it was observed that this sometimes even fails without error. In such a case mergedlines might only contain a single LineString (one street segment) instead of a MultiLineString (the combined network of all street segments). This completely messes up the following nearest_points().

Wrapping the argument in ``list()`` seems to be a valid workaround. It may come with a performance cost, as noted here:
https://github.com/geopandas/geopandas/issues/1820
https://github.com/geopandas/geopandas/issues/2171
This issue may disappear when shapely 2.0 is released (then pygeos is merged with shapely).

I have linked a PR that fixes the issue, with commit b25921f0536f9612b3e4f5f0817d785fd9c4fd99.

badf58030858c69363d4744ab046d8a2ef4855ed
The PR also addresses "UserWarning: CRS not set for some of the concatenation inputs." that can show up with geopandas 1.12.x. It is fixed by explicitly setting the crs on all concat inputs.

e571d35b108cf321dbe8392eb8a7458a064fa5e0
This also implements a small convenience function check_duplicate_geometries() within process_geometry(), for testing the input GeoDataFrame for duplicate geometries and plotting them. These can cause nasty errors and can be hard to find.

### LINKED ISSUE #108: Forcing an index reset on the building data is too restrictive
Our building data may have a special index, for example building addresses could define the index for more transparency elsewhere. Forcing a reset of this index in ``process_geometry()`` can cause inconveniences.

Specifically, I have a case where I want to use settings["heat_demand"] = "series" instead of a single P_heat_max value column. For this to work, the indices of the building geometry and the timeseries column names need to match.
They did match in my case, until the index of the geometry data got reset by ``process_geometry()``. It is inconvenient to have to reset and synchronize both indexes beforehand.

Instead with commit 4b5560acef360818ba87eba2d749291b4c6a7dfa I propose a new argument 'reset_index : bool' to process_geometry(). Default is True. If set to False, the index is not reset.
Apart from simply not using .reset_index(), another important change was 
```
if not reset_index:
    # Connection lines are ordered the same as points. Match their indexes
    lines_consumers.index = consumers.index
    lines_producers.index = producers.index
```
Furthermore, everywhere a string like 'consumers-Hans-Straße 8' is later deconstructed to 'consumers' and 'Hans-Straße 8', ``.split('-', maxsplit=1)`` instead of ``.split('-')`` has to be used (which would yield 'Hans')

Even debugging can become easier, because now nodes can have names like 'consumers-ABC123' which may be easier to identify.

This can lead to errors if the index contains duplicates for some reason. But this is the user's responsibility if they set reset_index=False. A ValueError is thrown.

### LINKED ISSUE #109: Method 'midpoint' produces too long connection lines
If large buildings stand close to the street, the default method of running the building connection lines from the street to the centroid of the building polygon can lead to much too long connection lines. 

In the attached PR (commit c1870cb7894def12b901483b943a02b1ea108b5d), I propose a new method='boundary' for the functions ``process_geometry()`` and ``create_points_from_polygons()``.
In one case of mine the house connections lines are now only 60% of the previous total length.

The 'midpoint' method (using the centroid) must already have been run, generating the default connection lines from street to centroid. Then the following rules apply:

- If there is only one intersection between that line and the boundary of the building, this intersection point is used as the connection point instead (and the connection line is shortened accordingly).
- However, complex building shapes can produce multiple intersections. In this case, the intersection with the 'convex hull' of the building is used instead. This may result in connection points that do not touch an actual building wall, but it should still be an improvement compared to the 'midpoint' method.
- In case of no intersections with the building boundary (possible for e.g. U-shaped buildings), the original centroid is used.

All of these conditions appear in the attached plots. For reference, they also show the convex hull.

The placement of the code within ``process_geometry()`` may not be the most elegant, because it depends on the midpoint method and ``create_object_connections()`` having already run. But I guess it is efficient.

Figure 2022-12-13 152818
Figure 2022-12-13 152829
Figure 2022-12-13 152841

jnettels:
This PR is mainly for implementing a new feature: Add a new method 'boundary' for creating the building connection point at the boundary/wall of the building, instead of the centroid. This results in much more realistic grid lengths.

Besides this new feature, this PR deals with a couple of issues that are closely related within the code. Creating separate PRs for the individual issues seemed too much of a hassle. Each subsequent commit works on its own, though.

- e571d35b108cf321dbe8392eb8a7458a064fa5e0, b25921f0536f9612b3e4f5f0817d785fd9c4fd99 and badf58030858c69363d4744ab046d8a2ef4855ed close #107
- 4b5560acef360818ba87eba2d749291b4c6a7dfa closes #108 
- c1870cb7894def12b901483b943a02b1ea108b5d closes #109

I have not updated any documentation. But since the only(?) documentation of [process_geometry()](https://dhnx.readthedocs.io/en/latest/api.html#dhnx.gistools.connect_points.process_geometry). seems to be the comment on the function itself, that at least is covered by my PR.

### DIFF
--- a/dhnx/gistools/connect_points.py
+++ b/dhnx/gistools/connect_points.py
@@ -312,13 +312,123 @@ def create_points_from_polygons(gdf, method='midpoint'):
     if gdf['geometry'].values[0].type == 'Point':
         return gdf
 
-    if method == 'midpoint':
+    if method == 'midpoint' or method == 'boundary':
+        # (method 'boundary' is performed later and needs the centroid)
         gdf['geometry'] = gdf['geometry'].centroid
         return gdf
 
     raise ValueError(
-        'No other method than >midpoint< implemented!'
+        "No other methods than 'midpoint' and 'boundary' implemented!"
     )
+
+
+def run_point_method_boundary(
+        consumers_poly, consumers, producers_poly, producers,
+        lines_consumers, lines_producers):
+    """Run 'boundary' method for finding the building connection point.
+
+    The 'midpoint' method (using the centroid) must already have been run,
+    generating the default connection lines from street to centroid.
+
+    If there is only one intersection between that line and the boundary
+    of the building, this intersection point is used as the connection
+    point instead (and the connection line is shortened accordingly).
+
+    However, complex building shapes can produce multiple intersections. In
+    this case, the intersection with the 'convex hull' of the building is used
+    instead. This may result in connection points that do not touch an
+    actual building wall, but it should still be an improvement compared to
+    the 'midpoint' method.
+
+    In case of no intersections with the building boundary (possible for e.g.
+    U-shaped buildings), the original centroid is used.
+
+    Parameters
+    ----------
+    consumers_poly : geopandas.GeoDataFrame
+        Polygons of the consumer buildings. Point geometries are also allowed,
+        but they are not changed.
+    consumers : geopandas.GeoDataFrame
+        Points of the consumer buildings (as returned by 'midpoint' method).
+    producers_poly : geopandas.GeoDataFrame
+        Polygons of the producer buildings. Point geometries are also allowed,
+        but they are not changed.
+    producers : geopandas.GeoDataFrame
+        Points of the producer buildings (as returned by 'midpoint' method).
+    lines_consumers : geopandas.GeoDataFrame
+        Connection lines from street to each consumer point.
+    lines_producers : geopandas.GeoDataFrame
+        Connection lines from street to each producer point.
+
+    Returns
+    -------
+    consumers : geopandas.GeoDataFrame
+        Updated points of the consumer buildings.
+    producers : geopandas.GeoDataFrame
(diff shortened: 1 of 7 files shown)


jnettels:
I feel like the failed check in the docs should not be my fault...

joroeder (approved):
Looks fine! I also checked an the import_osm_example with `method=boundary` and it works well.

When there is time, we should start writing tests also for the geometry processing part, and also add some infos to the documentation. But as this is still open for all the geometry processing part, this should not delay this PR.

PR #116, opened 2023-03-29
Update to oemof-solph v0.5.x

joroeder:
* Describe your pull request as transparent as possible
  * Update to oemof-solph v0.5.0
  * Fix deprecations warnings


joroeder:
The tests succeed also with python 3.10, but somehow the coverage report does not work. This will be fixed in a separate PR.

PR #122, opened 2023-04-19
Fix docs build on readthedocs

joroeder:
* see title and readthedocs build log


joroeder:
Okay, seems to work now.

PR #124, opened 2023-04-20
Add long tutorial for the DHNx pandapipes workflow

joroeder:
see title


joroeder:
@uvchik @jnettels If somebody has time to have a look at the tutorial, feedback is welcome!

If you have no time, no worries, let me know and I will merge it the next days ...

examples/tutorial_dhnx_pandapipes/tutorial_dhnx_pandapipes.py
```diff
@@ -0,0 +1,788 @@
...
+
+Contributors:
+
+- Joris Zimmermann
+- Johannes Röder
+
+"""
+import logging
+import math
+
```
jnettels, line 51:
```suggestion
import os
```

examples/tutorial_dhnx_pandapipes/tutorial_dhnx_pandapipes.py
```diff
@@ -0,0 +1,788 @@
...
+    tol_v=1e-3,
+)
+
+print(pp_net.res_junction.head(n=8))
+print(pp_net.res_pipe.head(n=8))
+
+# # Example of exports of the results
+
+# to excel
+with pd.ExcelWriter('results/results_fine.xlsx') as writer:
```
jnettels, line 730:
```suggestion
filepath = 'results/results_fine.xlsx'
if not os.path.exists(os.path.dirname(filepath)):
    os.makedirs(os.path.dirname(filepath))
with pd.ExcelWriter(filepath) as writer:
```
The 'results' folder does not exist when the script runs for the first time, so we need to create it here

jnettels (commented):
This is a great addition to dhnx.
As noted with the "suggestion" feature, a folder creation is missing from the script.
Also I got a weird ``fiona`` error which was fixed by rearranging the order of the import to this.
```
import logging
import math
import numpy as np
import matplotlib.pyplot as plt
from shapely import geometry
from CoolProp.CoolProp import PropsSI
import networkx as nx
from oemof.tools import logger
import pandas as pd
import pandapipes as pp
import osmnx as ox
```
But maybe that is just me, because in that environment I was still running shapely 1.8 and stuff is weird right now with the switch to shapely 2.0 and how all the other packages handle it.

As a more general note, it would be great to not only have this example script for pandapipes, but actually have a interface function which translates an dhnx network to the required input for pandapipes. That would be much cleaner and more reliable, hopefully

PR #141, opened 2024-10-09
Fix/solph network 05

p-snft:
Adapt to the following changes in packages this depends on:

* Constructor `oemof.network.Node` does not accept unknown keyword arguments anymore (stable release v0.5.0).
* Investments in `oemof.solph` have a period index (to allow for investments at different points in time).

* Also, the code style is changed to Black (as used in solph and network).


p-snft:
I just realised that you already addressed the issues (except the code style) in #134 and #137.

PR #145, opened 2025-01-17
Fix workflow issues

### LINKED ISSUE #146: Workflows are currently broken
The workflows of this repository are currently broken.

The workflow lint.yaml fails because ```actions/upload-artifact@v1``` is deprecated.
Also, the current v4 requires all uploaded artifacts to have unique names, but currently the python 3.8 and 3.10 job create the same artifact name.

In lint.yaml ```actions/setup-python@v1``` fails because the selected python version 3.8 is deprecated.

I have created a PR to fix those issues.

jnettels:
This PR is used to find fixes for the current workflow issues.

Fixes #146

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,9 +20,9 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.13
 
       - name: Install Python dependencies
         run: pip install flake8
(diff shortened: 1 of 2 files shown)


jnettels:
My solution requires the uploaded artifacts to have unique names.
The old artifact "tox-gh-actions-dist-python" has now two versions "tox-gh-actions-dist-python3.8" and "tox-gh-actions-dist-python3.10". The new names could cause problems if those artifact are used anywhere else. Does someone know?

p-snft (approved):
For other packages (e.g. solph) we upload the artifact only for one version. IMHO, this solution looks cleaner. I would suggest to test it here in DHNx.

PR #157, opened 2025-12-02
Feature/improved welding

p-snft:
The current graph pre-processing code has the problem that it keeps detours if there are forks on the shorter path. Further, the current "welding" code is slow. This PR is to tackle both, using `networkx` as a vehicle for efficient graph algorithms.


jnettels:
Hi, thanks for all the hard work updating the project!
I just noticed this PR and want to compare it to the old welding, which really is very slow. I hope I find the time to test it. Will you let me know when it is ready to test?

Just from a quick glance, it seems like the new ``go.drop_detours(lines_all)`` may have the same purpose as the existing ``lines = go.drop_parallel_lines(lines)``. The old one does not need a for-loop. I wonder which one is faster.

p-snft:
> Just from a quick glance, it seems like the new `go.drop_detours(lines_all)` may have the same purpose as the existing `lines = go.drop_parallel_lines(lines)`. The old one does not need a for-loop. I wonder which one is faster.

The old approach just deleted strictly parallel lines, i.e. two nodes where connected by more than one edge without additional edges in between. The new one removes detours even if the shorter connection consists of multiple segments. Note that `drop_parallel_lines` fails here because the shorter connection has forks connecting consumers.

p-snft:
Here we go, the algorithm is now done including the stuff needed for plotting.

**Welding: 5 seconds**
welding

**Simplify: 0.35 seconds**
simplified

PR #163, opened 2026-06-23
Fix errors in building docs

jnettels:
The docs were not building for quite a while.

Fixed errors:
- Some errors about broken file paths resulted from moving the folder ./dhnx to ./src/dhnx
- One link caused ``415 Client Error: Unsupported Media Type for url``, despite working in a browser (https://www.schweizer-fn.de/stroemung/rauhigkeit/rauhigkeit.php). Check is now ignored.

Fixed warnings:
- The logs also show ``WARNING: LaTeX command 'latex' cannot be run (needed for math display), check the imgmath_latex setting`` and equations were indeed broken in my local test. Switching the equation rendering to MathJax solved it for me.
- ``WARNING: Pygments lexer name 'txt' is not known [misc.highlighting_failure]``: Change ``code-block:: txt`` to ``code-block:: text``

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,6 +54,8 @@ linkcheck_ignore = [r"https://requires.io/.*"] + (
     [
         r"https://github.com/oemof/DHNx/issues/*",
         r"https://github.com/oemof/DHNx/pulls/*",
+        # 'Unsupported Media Type' despite valid link in 2026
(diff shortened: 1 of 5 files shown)


p-snft (approved):
Thanks for fixing this. It looks like I introduced some of the problems long time ago in 0398951.

PR #164, opened 2026-06-24
Features/update existing pipes

jnettels:
## Summary
- [x] Fix errors when using "existing" pipes for an optimization
- [x] Update attributes of returned pipe network
- [x] Make new ``simplify`` approach compatible (wait for #166)
- [x] Update documentation
- [x] Update changelog
- [x] Do some more tests in my own projects (marked as draft until then)

## Problem / Motivation
dhnx has support for mixing new and existing pipes in the optimization.

In the current state, dhnx does not return existing pipe segments with the results of the optimization process. This is somewhat logical, if we ask *"Dear dhnx, what new pipes do I need to invest in?"*

A user can work out their own solution for combining the existing pipes from their input data with the new pipes from the dhnx results. However, to me, this feels very unintuitive. My question would be *"Dear dhnx, what does the network look like that supplies heat to all the consumers?"*

Furthermore, some information is lost along the way, specifically:
- What actual thermal power flow (kW) is required in each pipe segment, including the existing pipes?
- Which of the existing pipe segments are actually needed?
- What heat losses occur in the existing pipe segments?
- What is the flow direction in each existing pipe segment?
  
I think dhnx should provide a clean way to get this information, instead of forcing each user to understand and pick apart the oemof results. I found ways to retrieve all those information for the oemof results, but I was unsure how to store them in
``network.results.optimization['components']['pipes']``

My first instinct was to write a solution without breaking changes, where I only add new attributes
- "flow"
    - for new pipes, we know: "capacity" = "flow" + "losses"
    - for existing pipes, we only know: capacity >= flow + losses
- "capacity_existing"
    - This is set to ``status_nominal`` from oemof.
    - If the pipe was used, this is just the capacity the user used in the input data
    - In case an existing pipe is unused, this becomes 0.
- "direction_existing"
    - The flow direction is an optimization result

This would avoid changing the columns "capacity" and "direction" which can be interpreted as "what is the newly installed pipe capacity"
https://dhnx.readthedocs.io/en/stable/optimization_models.html#results

This prevents breaking changes, but feels less clean. What I now did instead:

## Solution
- Add a new dhnx boolean setting "return_existing"
  - If False, keep columns as before
  - If True
    - "capacity" and "direction" are updated.
        - For existing pipes, "capacity" equals the input value if pipe is used, otherwise 0
    - "existing" is kept in the results
- Add the new column "flow" in any case, because for "existing pipe"-scenarios it is crucial and for new pipes it is nice to have.

This should avoid breaking changes as far as I can see.

## Other noteworthy changes
- d0ffb08183267d70d26533e06ce44c539dbf119e Replace the deprecated ``nominal_value`` with ``nominal_capacity`` 
- bbccf71b28e97c9368d6fb17fadeafe340f62677 Run a ``fillna({"existing": 0}`` if that column is used in the input data. I ran into issues when I only defined existing=1 for those pipes and left the new pipes NaN.
- 2575bcb1c731b213fb9d94664ccc1d9e00e70436 Use a more explicit condition ``if q["existing"] in [1]:`` instead of ``if q["existing"]:`` because it evaluates as ``True`` for ``NaN``. I guess it was related to the point above.
- a4b7dfe30dfb0cc62dc8e4fe9eed08fac53519d5 When the setting ``bidirectional_pipes`` is False, new heatpipes are created in both directions, to allow potential flows in both directions. When trying to use existing pipes, I ran into errors like this
  ``
  ValueError: No value for uninitialized VarData object InvestmentFlowBlock.invest[infrastructure_heat_bus_forks-67,infrastructure_heat_pipe-generic_forks-67-forks-75,0]
  ``
  My network became solvable once I added the same logic for exising pipes.  
- 3c0d9ee42fac7dd76a0537c30fc57dd22ecb04ee ``weld_segments()`` had to be updated, to be compatible with existing pipes.
  If there are existing pipes in the input data, we do not want to weld/merge/dissolve pipe segments with different capacities, which is what happened before. There is a new attribute ``retain_unique_values=['capacity']`` that makes sure unique values like 'capacity' are preserved.
  - Superseded by ``simplify()``
- d5d8dba393e5a4e1ea4176bb22c31bd0646fc8b9 Always derive flow direction from flow attribute. Use a consistent approach for new and existing pipes instead of using 'size' attribute, which yields inconsistencies for existing pipes and cases with 'bidirectional_pipes'=True.
- 8129f985467b6f4f2d64ead2316a5ef8f74a15c6 Add a ``FutureWarning`` for the switch from ``welding`` to ``simplify``


jnettels:
I would delegate a new test of a complete solver workflow with cbc to a new PR. That should improve coverage massively.

src/dhnx/optimization/dhs_nodes.py
```diff
@@ -75,7 +74,7 @@ def add_nodes_dhs(opti_network, gd, nodes, busd):
         d_labels["l_1"] = "infrastructure"
         d_labels["l_2"] = "heat"
 
-        if q["existing"]:
+        if q["existing"] in [1]:
```
p-snft: Do you mean the following?

```suggestion
        if q["existing"] == 1:
```

p-snft (commented):
I have mixed feeling about this. The PR improves functionality, but even explicitly increases technical debt.

> I would delegate a new test of a complete solver workflow with cbc to a new PR.

I cannot really blame you, as the architecture of dhnx isn't really modular, so (unit) testing also is an issue. At the moment, I feel like we need a big refactor anyway, but I do not see who has dedicated time for that purpose.

src/dhnx/optimization/dhs_nodes.py
jnettels: Yeah, at some point I though I needed to test ``q["existing"] in [1, True]``, but that was not necessary. I now changed that in ac30fef

jnettels:
I realized cbc was already installed in the workflows, so implementing new tests for the complete optimization process was quite simple, after all. I now included two new tests that cover several different options. Now we just crossed 70% total coverage. I feel like increasing the coverage further is beyond the scope of this PR.

p-snft:
The test coverage does not really say anything, as we don't have unit tests here. It's only that the integration test touches 70 % of the code. (This will not help to find out where in the code stuff is broken if the test fails.)

As I think this PR is a clear improvement, I will merge. Thanks.

PR #181, opened 2026-08-12
Implement pressure loss calculation via pandapipes

### LINKED ISSUE #162: Used friction factor formula provides wrong values
I am under the impression that the used formula to calculate the friction factor for pressure loss in pipes as proposed by Bordin is wrong.

Comparing the results of Bordin's formula with those given by Colebrook-White for different pipe diameters, flows and surface roughnesses shows that Bordin's formula is off by a factor as high as 2.

It is probably best to implement the Colebrook-White equation or some other more well-known approximation without iteration.

jnettels:
I have added a new argument ``calculation`` to ``delta_p()`` and the functions which call it (``v_max_secant()``, ``v_max_bisection()``) which allows using a ``pandapipes`` simulation to calculate pressure losses in a pipe as an alternative to the existing internal implementation.

* [x] Update code
* [x] Add new test
* [x] Update changelog
* ~Update documentation~

Fixes #162

## Context
dhnx includes the ``precalc_hydraulic`` module which allows converting the resulting capacity of each pipe into norm diameters.
For this, a pressure loss calculation is needed, for which ``precalc_hydraulic`` provides an internal solution.

In my workflows, I design a network with dhnx and then run a simulation of temperature and pressure distribution with pandapipes to check if the network is valid in terms of pressure drops and velocities. In some rare edge cases I noticed differences where dhnx was supposed to only give e.g. a maximum of 100 Pa/m pressure losses in any given pipe segment but pandapipes showed slightly larger values.

Implementing a new function ``delta_p_pandapipes()`` in ``precalc_hydraulic`` solved the issue because it synchronizes the precalculation and post-processing. It allows calculating the pressure loss with pandapipes instead of using the internal implementation.

Independently, in #162 another user describes issues with the internal implementation. Instead of trying to fix/improve the internal pressure loss equations I think it is the best approach to use external modules that have already put more resources into this problem. (Nevertheless: pandapipes itself offers different friction models, and judging by the issues there, not all of them are perfect, either.)

## Considerations / feedback

* My approach gives room to add other calculation methods in the future, which may coexist. I hope that is useful
* Running ``pandapipes`` is significantly slower than the internal method. But in the context of a long optimization procedure that is fine
* ``precalc_hydraulic`` is briefly mentioned in the documentation, but not actually documented. Thus there was no documentation to update
* I have added matching tests
* In order for the new test to work, I have added ``pandapipes`` to the deps in testenv in ``tox.ini``. But I am unsure if this is the best way and I need feedback. I guess I could create a new entry of optional dependencies in ``pyproject.toml`` instead, but that feels like overkill to me. If a user knows they want to use pandapipes alongside dhnx, they can just install it.
* ``pandapipes`` requires ``pandapower``, which is currently still restricted to ``pandas<3.0``. Thus with this change all of our tests here will run with ``pandas 2.x``, which is not ideal, I think. But I see no way around that. Thoughts?

### DIFF
--- a/docs/whatsnew/v0-1-0.rst
+++ b/docs/whatsnew/v0-1-0.rst
@@ -40,6 +40,16 @@ API changes
     found, there will be a warning.
 
 
+* Implemented pressure loss calculation via pandapipes
+
+  * When calling ``delta_p()``, ``v_max_secant()`` and ``v_max_bisection()``
+    from the ``precalc_hydraulic`` module, users can now choose pandapipes
+    as the calculation method for the pressure loss of a pipe. This is done
+    by setting ``calculation="pandapipes"``. Additional ``kwargs`` like
+    ``friction_model`` can be passed to the pandapipes calculation.
+    Requires installing ``pip install pandapipes`` separately. The default
+    setting is the existing ``calculation="internal"``.
+
 New features
 ^^^^^^^^^^^^^^^^^^^^
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,7 @@ Changelog = "https://dhnx.readthedocs.io/en/latest/whats_new.html"
 ]
 "hydraulic" = [
     "CoolProp",
+    "pandapipes",
 ]
 
 [tool.black]
--- a/src/dhnx/optimization/precalc_hydraulic.py
+++ b/src/dhnx/optimization/precalc_hydraulic.py
@@ -331,6 +331,82 @@ def delta_p(
     pressure=101325,
     R_crit=2320,
     fluid="IF97::Water",
+    calculation="internal",
+    **kwargs,
+):
+    r"""
+    Function to calculate the pressure loss in a pipeline
+
+    Parameters
+    ----------
+    v : numeric
+        :math:`v`: flow velocity [m/s]
+
+    d_i : numeric
+        :math:`d_i`: inner pipe diameter [m]
+
+    k : numeric
+        :math:`k`: roughness of inner pipeline surface [mm]
+
+    T_medium : numeric
+        :math:`T_{medium}`: fluid temperature [°C]
+
+    length : numeric
+        :math:`l`: length of the pipe [m]
+
+    pressure : numeric
+        :math:`p`: pressure in the pipe [Pa]
+
+    R_crit : numeric
+        :math:`Re_{crit}`: critical Reynolds number between laminar and
+        turbulent flow [-]. Only used for ``calculation="internal"``.
+
+    fluid : str
+        name of the fluid used
+
+    calculation: str
+        Type of calculation, default: "internal". If "pandapipes" is chosen,
+        the pandapipes module is used for the calculation of the pressure drop.
+
+    **kwargs: dict
+        Additional keyword arguments for the selected external calculation.
+        For pandapipes, e.g. 'friction_model' can be specified.
+
+    Returns
+    -------
+    Pressure drop [Pa] : numeric
+
+    """
+    if calculation == "internal":
+        d_p = delta_p_internal(
+            v, d_i, k, T_medium, length, pressure, R_crit, fluid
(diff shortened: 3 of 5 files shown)


jnettels:
Hi @p-snft , if you have the time I would very much appreciate your feedback, especially on how to handle pandapipes as a dependency in ``tox.ini``.
This is the rest of my currently unmerged changes to dhnx. Maybe all of our recent changes are worth a new release?

p-snft (commented):
> My approach gives room to add other calculation methods in the future, which may coexist. I hope that is useful.

My vision anyway is to develop DHNx to a part of a more generic workflow. So, instead of having just more methods we might have a data format and separate scripts to process it. (I would wait for oemof datapackage, here.)

An additional thought: Would it make sense to replace the old (internal) calculation?

> Running pandapipes is significantly slower than the internal method. But in the context of a long optimization procedure that is fine

Fine by me.

> In order for the new test to work, I have added pandapipes to the deps in testenv in tox.ini. But I am unsure if this is the best way and I need feedback. I guess I could create a new entry of optional dependencies in pyproject.toml instead, but that feels like overkill to me. If a user knows they want to use pandapipes alongside dhnx, they can just install it.

I will add that, as it's cleaner.

> pandapipes requires pandapower, which is currently still restricted to pandas<3.0. Thus with this change all of our tests here will run with pandas 2.x, which is not ideal, I think. But I see no way around that. Thoughts?

This has to be fixed upstream but I won't see it as a blocker here.

jnettels:
> An additional thought: Would it make sense to replace the old (internal) calculation?

Although there is the one bug report #162, in my projects the differences between ``internal`` and ``pandapipes`` are very small. So I think the internal calculation is a worthy default in case you cannot or do not want to install pandapipes. I would not want to only rely only on that.

(I now had to create a fork that allows using pandas 3, because pandas 2 just caused too many issues in other packages that have moved on... https://github.com/e2nIEE/pandapipes/issues/821)

> My vision anyway is to develop DHNx to a part of a more generic workflow. So, instead of having just more methods we might have a data format and separate scripts to process it. (I would wait for oemof datapackage, here.)

Depending on what you are actually after, I may have created such a generic workflow. I think I have mentioned it before, but I have not heavily advertised it:
https://github.com/jnettels/dhnx_addons/
And more specifically this example method:
https://github.com/jnettels/dhnx_addons/blob/main/src/dhnx_addons/dhnx_addons.py#L302

- It downloads building geometry and street data from openstreetmap
- Applies methods to determine annual heat demand of each building
- Runs ``lpagg`` (which internally runs ``oemof.demand`` plus a few extra steps) to create load profiles for all buildings
- Peak thermal power of the load profiles defines the ``P_heat_max`` of each building required for ``dhnx`` (although the load profiles and a time series optimization are also possible)
- Runs ``dhnx`` for the given network and buildings. Does stuff like repeating the optimization if the result contains loops, which are almost always just a result of limited optimization time
- Has a complex post-processing step that applies a certain simultaneity logic to the ``dhnx`` result
- Runs a wrapper around ``pandapipes`` that allows simulating temperature and pressure distribution in the network given by ``dhnx``

I am proud of what it does. But it does so much that it is hard to document everything. Plus the idea is that it is just a collection of various useful functions, from which you always pick the ones you need. It breaks easily due to the many dependencies it has. Also it has grown very organically, which is to say that I am not proud of how the files and functions are structured. And the name ``dhnx_addons`` is stupid, but that was what it started as, just of few more extras on top of dhnx.

My takeaway is to expand dhnx only where absolutely necessary, to keep the scope of this individual project as clean and focused as possible. It would be nice to move ``dhnx_addons`` into something more stable and polished, but I can only muster the development time for the stuff that I actually need to work.

Still, maybe it is something that is worth knowing about.

---
Anyway, you can merge this PR now if you want to. Thanks!

---
Thoughts on releasing a new version? Although #183 just came in and seems like a reasonable request

PR #184, opened 2026-08-21
Replace optional dependency warning prints

### LINKED ISSUE #183: Potentially unneeded notifications about optional dependencies
When importing dhnx there are notifications in the form "Need to install ... to ..." about optional dependencies regardless if they are actually needed. We are launching many python processes collecting output to process later on ... and want to keep this as "clean" as possible, so this messages are distractive and therefore not wanted.
Please take care that this messages are only printed out if they are really helpful ;-)

p-snft:
The idea behind the OptionalDependencyPlaceholder is that it raises an Error only once code asks for an attribute. To work with this approach, function and class imports had to be replaced by module imports.

### DIFF
--- a/src/dhnx/helpers.py
+++ b/src/dhnx/helpers.py
@@ -20,3 +20,18 @@ def sum_ignore_none(*items):
         sum_ignoring_none = None
 
     return sum_ignoring_none
+
+
+class OptionalDependencyPlaceholder:
+    def __init__(
+        self,
+        module_name: str,
+        functionality: str,
+    ):
+        self._functionality = functionality
+        self._module_name = module_name
+
+    def __getattr__(self, name):
+        raise ModuleNotFoundError(
+            f"Need to install {self._module_name} to {self._functionality}."
+        )
--- a/src/dhnx/helpers.py
+++ b/src/dhnx/helpers.py
@@ -26,6 +26,7 @@ def sum_ignore_none(*items):
 
 
 class OptionalDependencyPlaceholder:
+    """Class to delay ImportErrors for optional dependencies"""
     def __init__(
         self,
         module_name: str,
@@ -37,7 +38,7 @@ class OptionalDependencyPlaceholder:
         self._original_error = original_error
 
     def __getattr__(self, _):
-        raise ModuleNotFoundError(
+        raise ImportError(
             f"Need {self._module_name} to {self._functionality},"
             + f" but {self._original_error}."
         )
@@ -47,6 +48,11 @@ def import_optional_dependency(
     name: str,
     functionality: str,
 ) -> types.ModuleType:
+    """Import wrapper for optional dependencies.
+
+    If possible, it will just import the module.
+    Otherwise, it returns a placeholder so that the ImportError is not risen
+    on import but only if the optional dependency is acutally used."""
     try:
         module = importlib.import_module(name)
     except ImportError as err:
--- a/src/dhnx/helpers.py
+++ b/src/dhnx/helpers.py
@@ -1,3 +1,6 @@
+import importlib
+import types
+
 import addict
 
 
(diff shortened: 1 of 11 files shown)


src/dhnx/helpers.py
```diff
@@ -20,3 +20,18 @@
...
+class OptionalDependencyPlaceholder:
+    def __init__(
+        self,
+        module_name: str,
+        functionality: str,
+    ):
+        self._functionality = functionality
+        self._module_name = module_name
+
+    def __getattr__(self, name):
```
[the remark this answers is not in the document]
p-snft: This is on purpose. The sense of this class is to delay `ModuleNotFoundError`s for optional dependencies.

jnettels:
I think I get the idea of how this is supposed to work and it looks quite clever. However I feel like this can cause bad "unforseen consequences" (but maybe this is just a lack of experience).

I looked up how pandas handles this and the gist is that imports of optional dependencies are not performed at the top of modules, but instead only within the functions where they are needed. They are, however wrapped in an ``import_optional_dependency()`` function that checks if the module is imported successfully and provides helpful install instructions in case they are not. It also checks for outdated versions, which is very nice.
For us, with e.g. ``geopandas`` this would mean duplicate imports in several places, but from my understanding this is not an issue in python (apart from extra lines of code).

Definition of import function: https://github.com/pandas-dev/pandas/blob/main/pandas/compat/_optional.py

Example usage: https://github.com/pandas-dev/pandas/blob/6c0ceb0eddfe9d6cb8c6a7c6679b2cbfc811b825/pandas/io/excel/_util.py#L84

If I had to do it, I would feel much more confident with this approach.
At the same time, I don't want to stop you if you think you can make your approach work.

One way or the other, getting rid of uncontrolled ``print()`` statements has my full support, because it annoys me, too :-)

p-snft:
The pandas solution has some advantages:
* It checks minimum versions. This could be added to my per-module approach. (But I think it is okay to just rely on error handling in the imported modules.)
* It allows to move the try/except blocks to one central place. For them, this means that they have to `import_optional_dependency()` at runtime, as this function directly triggers the warning. I like the aspect of removing duplication but would like to keep the late failure.
* As it import at runtime, it allows to implement fallback code paths if an optional dependency is not present. (We do not have fallback code, so this does not matter for us - at least not yet.)

My approach should be easier if you have functions that just fail if the optional dependency is not present. After the import, you just assume your optional dependency is present. The customised import handling is just meant to give an easy to understand error message. As we currently let users select code paths and have no default/better code path plus an automatic (but possibly worse) fallback, I think we do not really profit from the pandas way to do it.

Update: I changed to something inspired by pandas that still allows late failure in combination with module-wide imports.

src/dhnx/optimization/precalc_hydraulic.py
```diff
@@ -25,14 +25,11 @@
...
+from dhnx.helpers import import_optional_dependency
 
-except ImportError:
-    print(
-        "Need to install CoolProp to use the hydraulic "
-        "pre-calculation module."
-    )
+CoolProp = import_optional_dependency(
+    "CoolProp.CoolProp", "use the hydraulic pre-calculation module"
+)
```
jnettels, line 32:
```suggestion
)
pp = import_optional_dependency(
    "pandapipes", "use pandapipes pressure loss calculations"
)
```

jnettels: This is another optional dependency that we can cover.
The old import statement further below then needs to be removed.

src/dhnx/optimization/precalc_hydraulic.py
jnettels:
```suggestion
```

jnettels (changes requested):

jnettels:
Thanks for the effort. As long as it works, this looks quite nice! Apart from the ``pandapipes`` import I do not think I have any further comments. At this point I hope we can trust the tests enough, I did not test it locally.

p-snft:
Thanks @jnettels for spotting the missed import. If you approve, I'll merge.

jnettels (dismissed):
Ok!
PS: I did already approve... but now I wonder if there should be at least one test for the missing import...? Sorry

This seems to work, e.g. in ``test_errors.py``

```python
from dhnx.helpers import import_optional_dependency

def test_import_optional_dependency():
    dependency = import_optional_dependency(
        "missing_optional_dependency", "run optional feature"
    )

    with pytest.raises(
        ImportError,
        match="Need missing_optional_dependency to run optional feature",
    ):
        dependency.some_attribute
```

jnettels (approved):
Awesome

